PR #49, opened 2014-06-11
Fix the install_requires field for the acme.motd plugin egg

jwiggins:
Renaming EnvisageCore -> Envisage caused the `MOTD_Using_Eggs` example to break.


jwiggins:
An oldie but goodie...

PR #48, opened 2014-05-28
Python 3 compatibility and Travis integration for it.

pradyunsg:
This PR is for a Port of Envisage to Python 3 using the same, common code-base for Python 2 and Python 3.

Major Changes:
  - `envisage.safeweakref`: uses a backport of `weakref.WeakMethod` from Python 3.3 for methods.
  - `envisage.developer.code_browser`: uses `ast` module instead of compiler package.
  - `envisage.plugins.remote_editor.communication.util`: handles `bytes` objects in Python 3 correctly.

Other than that for Python 3, Handling of imports of removed or renamed standard library modules has been added. Also, relative imports have been used for in-package imports.

PR #58, opened 2015-08-20
Port to Python3.

prabhuramachandran:
This is primarily using 2to3 while using a little bit of six.  There are
several manual changes.  These changes should allow us to use the same
code base with Python 2 and 3.  All tests except 2 pass on Python 3.4.  These two use the old compiler module and need a bit more work.  If six is not desired, we can do that with a bit of work.


prabhuramachandran:
Closing this for now as Pradyun's PR #48 actually did the work earlier.  I didn't add his remote correctly and didn't realize that it was in good shape.  I'll use this commits (while removing the travis stuff he added) and send a fresh PR based on his commit so it works with master.

PR #60, opened 2015-09-03
Fix the travis-ci setup

itziakos:
This PR updates the travis-ci setup.

in more details
- ~~Use travis-ci containers~~ Have to use sudo to remove the old distribute package that was causing problems
- Run tests on python 2.6 (before no tests were running on python 2.6)
- Run all the tests (before we were running only the envisage/plugin tests)
- skip tests when GUI toolkit is not available
- skip tests when ipython is not available
- force ipython 3.2.1 (it looks that there issues with ipython 4)
- Update travis-ci dependencies 
- Add coverage reporting


itziakos:
So far I have remove most of the errors, except from the pkg_resources issue. This is the last to address.

itziakos:
It was that simple just do not use containers and remove distribute from the system python (just a blast).

@pberkes, can you please have a look at this PR.

.travis.yml
```diff
@@ -1,20 +1,32 @@
...
   - python setup.py develop
 script:
-  - if [[ $TRAVIS_PYTHON_VERSION == *_with_system_site_packages ]]; then nosetests envisage; else nosetests envisage/tests; fi
-notifications:
-  email:
-    - travis-ci@enthought.com
+  - coverage run -m nose.core envisage -v
+after_success:
+  - pip install codecov
+  - codecov
```
pberkes: nit: Missing newline

envisage/plugins/ipython_kernel/tests/test_internal_ipkernel.py
```diff
@@ -1,7 +1,12 @@
 import unittest
 
-from IPython.kernel.zmq.kernelapp import IPKernelApp
+try:
+    import IPython  # noqa
+except ImportError:
+    from nose.plugins.skip import SkipTest
+    raise SkipTest('IPython not available')
```
pberkes: Is this still necessary? Can it be removed?

pberkes:
Ouch that looks like a lot of work. Thanks!
LGTM, I just have a concern about all the test skipping. Travis seems to run all of them, though.

envisage/plugins/ipython_kernel/tests/test_internal_ipkernel.py
itziakos: IPython is not part of the hard dependencies of envisage, so the right thing to do is to skip tests that require an optional depedency. Furthermore, it looks that IPython does not support python 2.6 (even thought the package is installed, the tests are skipped).

itziakos: see http://ipython.org 

```
IPython supports Python 2.7 and 3.3 or newer. Our older 1.x series supports Python 2.6 and 3.2.
```

.travis.yml
itziakos: Done

itziakos:
Thanks Pietro, I could not help myself, the previous setup was very bad. There is no point in setting up automatic testing and not keeping it up to date (or in this case not testing much :P)

pberkes:
LGTM

PR #67, opened 2016-03-04
Add tox for testing package install

kitchoi:
- Added tox settings
- Added MANIFEST.in
- Updated setup.py for package data
- Added __init__ for a few test packages
- Fix one local import for tests in py3

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -153,7 +153,12 @@ if __name__ == "__main__":
         install_requires = __requires__,
         license = "BSD",
         packages = find_packages(),
-        package_data = {'': ['images/*', '*.ini',]},
+        package_data = {'': ['images/*', '*.ini',],
+                        'envisage.tests': ['bad_eggs/*.egg',
+                                           'eggs/*.egg',
+                                           'plugins/pear/*.py',
+                                           'plugins/banana/*.py',
+                                           'plugins/orange/*.py']},
         platforms = ["Windows", "Linux", "Mac OS-X", "Unix", "Solaris"],
         zip_safe = False,
     )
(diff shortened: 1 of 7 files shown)


kitchoi:
@sjagoe can you take a look please?  The tarball on pypi at the moment has missing files and fail tests.  tox can ensure that doesn't happen again.  thanks @cournape for his help!

setup.py
```diff
@@ -153,7 +153,12 @@ def write_version_py(filename=_VERSION_FILENAME):
         install_requires = __requires__,
         license = "BSD",
         packages = find_packages(),
-        package_data = {'': ['images/*', '*.ini',]},
+        package_data = {'': ['images/*', '*.ini',],
```
sjagoe: I think this is going to include `tox.ini` in eggs, whereas I expect we just want it in the sdist.

sjagoe: Specifying it only in MANIFEST.in as `include tox.ini` may be a better approach.

sjagoe: ah, I see I was wrong actually. It does not include top-level ini files. Ignore the above comment.

sjagoe:
@kitchoi Please add a changelog entry

setup.py
kitchoi: I don't think tox.ini would be included because it is not inside the envisage package.

kitchoi:
@sjagoe I have updated changelog and I checked the egg problem with setuptools bdist_egg.  Please let me know if I should do something different.

sjagoe:
LGTM. I'll produce a 4.5.1 release.

PR #77, opened 2016-09-28
Create tasks application without gui

sjagoe:
@corranwebster 

This fixes some bad trait definitions from #76 as well as making it possible to create a `TasksApplication` including the `TasksPlugin` without creating a GUI.

### DIFF
--- a/envisage/ui/tasks/tasks_application.py
+++ b/envisage/ui/tasks/tasks_application.py
@@ -60,10 +60,10 @@ class TasksApplication(Application):
     task_extensions = ExtensionPoint(id=TASK_EXTENSIONS)
 
     # The list of task windows created by the application.
-    windows = List('envisage.ui.tasks.task_window.TaskWindow')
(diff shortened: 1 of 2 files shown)


corranwebster:
Generally looks fine, but haven't run it.

Test failure looks unrelated, but probably should be fixed sooner rather than later.

sjagoe:
Frankly, I don't even see how the Python 3 test run is supposed to work on travis. The rest of the file references `pip2` and debian package `python-qt4`, which is Python 2.7 only.

sjagoe:
The Python 3 issue seems to be caused by a change in traits at enthought/traits#115:

https://github.com/enthought/traits/blob/9e5741bf49943dd37a863d15cdaef32a32cd22cc/traits/trait_handlers.py#L2552
https://github.com/enthought/traits/blob/9e5741bf49943dd37a863d15cdaef32a32cd22cc/traits/trait_handlers.py#L2495

sjagoe:
I'll open a new PR for that.

sjagoe:
Python 3 test fix in #78

PR #90, opened 2017-03-17
Fix task layout serialization under Python3

pankajp:
In Python3, pickle serializes to bytes so file has to be opened
in binary mode.


corranwebster:
LGTM

PR #120, opened 2018-10-25
Install deps from source (github)

rahulporuri:
At the moment, this is specifically to install traits from source for testing.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,15 +8,16 @@ env:
 
 matrix:
   include:
-    - env: RUNTIME=2.7 TOOLKITS="null pyqt pyside wx"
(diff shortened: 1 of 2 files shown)


rahulporuri:
closing this PR as it was only meant to test envisage against traits master on CI.

rahulporuri:
the infinite recursion is happening in the call to `inspect.getmembers` here - https://github.com/enthought/envisage/blob/b0d6cf644c0c7495eed0d3d746e924c3bece7fcc/envisage/plugin.py#L368

rahulporuri:
never mind. #119 is the relevant issue

rahulporuri:
the other relevant issue here is #94

PR #142, opened 2019-03-06
String repr for ExtensionPoint objects

### LINKED ISSUE #141: String repr for ExtensionPoint instances
Currently `ExtensionPoint` reprs use the standard Python repr i.e. `'<envisage.extension_point.ExtensionPoint object at 0x7f125d4ce550>'`, which is not useful in anyway. It would be nice if `__str__` returns the `ExtensionPoint.id` attribute.

rahulporuri:
The str repr should display the `id` attr of `ExtensionPoint` instances.
Fixes #141

### DIFF
--- a/envisage/extension_point.py
+++ b/envisage/extension_point.py
@@ -141,6 +141,10 @@ class ExtensionPoint(TraitType):
 
         return
 
+    def __str__(self):
+        """ String representation of an ExtensionPoint object """
+        return self.id
+
     ###########################################################################
     # 'TraitType' interface.
     ###########################################################################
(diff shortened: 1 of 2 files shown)


envisage/extension_point.py
```diff
@@ -141,6 +141,12 @@ def __init__(self, trait_type=List, id=None, **metadata):
 
         return
 
+    def __repr__(self):
+        """ String representation of an ExtensionPoint object """
+        return "ExtensionPoint(id={})".format(self.id)
+
+    __str__ = __repr__
```
corranwebster: This is not needed.

corranwebster:
Minor nit, but otherwise LGTM.

envisage/extension_point.py
rahulporuri: addressed in commit https://github.com/enthought/envisage/pull/142/commits/280f6f5b063ae664cf9d4ad48dd76c606ebc220e

rahulporuri:
Thanks @corranwebster !

PR #192, opened 2019-09-12
Add license headers to all files

### LINKED ISSUE #24: License headers
It would be good if all source files could contain the proper license header in the file.  This greatly helps with license audits.

rahulporuri:
fixes #24 

Most files didn't have a license header to begin with but a few did, which was updated to the currently standard format for enthought oss packages.


rahulporuri:
@mdickinson are you okay with me rebasing this branch on top of the current master?

PR #296, opened 2020-08-10
Update copyright headers

rahulporuri:
- add new copyright_header flake8 plugin (stolen from traits)
- update copyright headers for 2020

NOTE : two modules don't have the enthought headers because they are based on IPython modules.


kitchoi:
I have not reviewed much at all.
Could you confirm please if the `copyright_header` package is going to get grabbed by envisage's `setup.py`?
The envisage setup.py calls `find_packages` within explicitly opting-in or opting-out packages.
I believe the copyright_header is not intended to be shipped with envisage.

rahulporuri:
>  Could you confirm please if the copyright_header package is going to get grabbed by envisage's setup.py?

Nice point. But from the looks of it, the package isn't grabbed by the envisage setup.py. Looks like that happens because `copyright_header` itself is not a python package but a top-level directory containing the package.

setup.cfg
```diff
@@ -1,4 +1,4 @@
 [flake8]
 exclude = examples, envisage/ui/single_project
 ignore = E266, W503
-per-file-ignores = */api.py:F401
+per-file-ignores = */api.py:F401, */heartbeat.py:H101, */kernelapp.py:H101, setup.py:H102,
```
aaronayres35, line 4: are these all specifically related to this PR?

rahulporuri: yes. H101 and H102 are the copyright header errors and heartbeat.py, kernelapp.py and setup.py are excluded because 

- `heartbeat.py` and `kernelapp.py` are borrowed/adapted from the IPython project. (I'm not sure if that means we can just add our copyright header on top of the IPython copyright header)
- `setup.py` generates a `version.py` file during installation the file has two copies of the copyright header.

rahulporuri: a total of 4 errors defined by `copyright_header.py` https://github.com/enthought/envisage/blob/26fd5196a587a14d209829159d61dd3df431565f/copyright_header/copyright_header.py#L97-L142

aaronayres35: I understand now

corranwebster: For stuff from IPython, if we have modified the files then we have copyright to the changes, so the copyright header should include both.  We want to be careful about not asserting a copyright date prior to our first modification, so 2005-2020 is probably not right.

corranwebster:
Not entirely happy with the copy-pasting of the copyright header checking code across the ETS codebases.

rahulporuri:
> Not entirely happy with the copy-pasting of the copyright header checking code across the ETS codebases.

@corranwebster yeah. I am also in favor of creating a separate/new repo/package (it might make sense to move this package to the github.com/enthought/ets repo). At the moment, it's only being used in traits and envisage. Once we update the headers in traitsui and pyface, we should be in a better position to create the standalone package.

rahulporuri:
closes #264

PR #317, opened 2020-10-20
Fix the attractors tasks application example

rahulporuri:
The fixes are mostly trivial - involving updating implicit imports to explicit imports.

To be able to run this example, you will need to install `numpy`, `scipy`, `mayavi` and `chaco` in the envisage devenv and then do `python -m attractors.run` from the `examples/plugins/tasks` directory.

### DIFF
--- a/examples/plugins/tasks/attractors/README.txt
+++ b/examples/plugins/tasks/attractors/README.txt
@@ -1,0 +1,12 @@
+Welcome to the attractors example.
+
+To run the application::
+
+  ``python -m attractors.run``
+
+from the ``tasks/`` directory.
+
+Note that this example application depends on the following additional
+packages::
+
+  numpy, scipy, chaco, mayavi
--- a/examples/plugins/tasks/attractors/attractors_application.py
+++ b/examples/plugins/tasks/attractors/attractors_application.py
@@ -4,7 +4,7 @@ from pyface.tasks.api import TaskWindowLayout
 from traits.api import Bool, Instance, List, Property
 
 # Local imports.
-from attractors_preferences import (
+from attractors.attractors_preferences import (
     AttractorsPreferences,
     AttractorsPreferencesPane,
 )
--- a/examples/plugins/tasks/index.rst
+++ b/examples/plugins/tasks/index.rst
@@ -4,7 +4,7 @@ To run the application::
 
   python -m attractors.run
 
-from the ``tasks/`` directory.
+from this directory.
 
 Note that this example application depends on the following additional
 packages::
--- a/examples/plugins/tasks/index.rst
+++ b/examples/plugins/tasks/index.rst
@@ -2,7 +2,7 @@ Welcome to the attractors example.
 
 To run the application::
 
-  ``python -m attractors.run``
+  python -m attractors.run
 
 from the ``tasks/`` directory.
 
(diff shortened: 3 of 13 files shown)


examples/plugins/tasks/attractors/README.txt
```diff
@@ -0,0 +1,12 @@
+Welcome to the attractors example.
+
+To run the application::
+
+  ``python -m attractors.run``
```
kitchoi, line 5: With `::` in the previous line, this line is interpreted to be in a code block such that the double backtick is not required (I think).

examples/plugins/tasks/attractors/README.txt
```diff
@@ -0,0 +1,12 @@
...
+
+from the ``tasks/`` directory.
```
kitchoi, line 7: Maybe we need a README in `examples/plugins/tasks` in addition to OR instead of this file. Otherwise one has to navigate into the `attractors` folder to read this file, and then change directory again to run the script from outside.

examples/plugins/tasks/attractors/README.txt
kitchoi: Should this file have an `.rst` file extension?
Not sure if this is to be exposed under etsdemo, in which case file can be named `index.rst` instead.

examples/plugins/tasks/attractors/attractors_application.py
```diff
@@ -4,7 +4,7 @@
 from traits.api import Bool, Instance, List, Property
 
 # Local imports.
-from attractors_preferences import (
+from attractors.attractors_preferences import (
```
kitchoi, line 7: This really does assume where `python` is run from, that's okay if this really isn't supposed to be run in any other way. `from .attractors_preferences import ...` would be free of that assumption in this case.

rahulporuri: i was on the fence about using implicit imports. updating them.

examples/plugins/tasks/attractors/README.txt
rahulporuri: that makes sense. and i'll rename this to `index.rst` because this is one of the examples I would like shipped with the package.

examples/plugins/tasks/attractors/attractors_application.py
rahulporuri: hmm. that did not work as expected. trying to run `python run.py` gives the error

```
  File "run.py", line 7, in <module>
    from .attractors_plugin import AttractorsPlugin
ModuleNotFoundError: No module named '__main__.attractors_plugin'; '__main__' is not a package
```

which makes sense because python does not know about the `attractors` package because it is not on path when i cd into the `attractors` directory.

kitchoi: OK... I am not sure how this is going to work under etsdemo if the examples need to assume certain python paths. Probably a separate issue.

examples/plugins/tasks/index.rst
```diff
@@ -0,0 +1,12 @@
+Welcome to the attractors example.
+
+To run the application::
+
+  python -m attractors.run
+
+from the ``tasks/`` directory.
```
kitchoi, line 7:
```suggestion
from this directory.
```

kitchoi (approved):
For future reference, relative imports would have enabled running the plugin from an outer directory, e.g. in `examples/plugins` folder:
```
python -m tasks.attractors.run
```
Or even from the source root directory:
```
python -m examples.plugins.tasks.attractors.run
```
In any case, running `run.py` directly from the `examples/plugins/tasks/attractors` won't work whichever relative or absolute import we use, unless we change with the python path somehow.

We don't seem to need that flexibility for now, so this is okay.

PR #326, opened 2020-10-20
TST: Switch on default warning flag for CI test command

### LINKED ISSUE #311: Switch on the default warning flag when running testsuite
Deprecation warnings triggered during test runs are visible, but deprecation warnings triggered prior to when a unittest test case is loaded are by default hidden from the console.

With that, deprecation warnings from trait types definition do not show up from running the current test command, because the warnings occur at import time.

See PR https://github.com/enthought/traitsui/pull/1059 for a solution to this.

aaronayres35:
fixes #311 

This PR makes the exact same fix that was done in https://github.com/enthought/traitsui/pull/1059.  It adds the adds the `-W default` flag.

### DIFF
--- a/etstool.py
+++ b/etstool.py
@@ -304,7 +304,7 @@ def test(edm, runtime, toolkit, environment):
     environ["PYTHONUNBUFFERED"] = "1"
     commands = [
         (
-            "{edm} run -e {environment} -- "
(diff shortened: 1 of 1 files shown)


rahulporuri (approved):
LGTM

PR #328, opened 2020-10-21
Trivial flake8 fixes for envisage.ui.single_project

kitchoi:
These are all the trivial flake8 fixes I have done before I realized that the package may be scheduled for removal (see #266).

They are pretty easy to do.  If the flake8 fixes don't go in, we might still want to update `setup.cfg` to reference #269 anyway.

### DIFF
--- a/envisage/ui/single_project/action/api.py
+++ b/envisage/ui/single_project/action/api.py
@@ -2,9 +2,10 @@
 # All rights reserved.
 #
 # This software is provided without warranty under the terms of the BSD
-# license included in LICENSE.txt and may be redistributed only
-# under the conditions described in the aforementioned license.  The license
+# license included in LICENSE.txt and may be redistributed only under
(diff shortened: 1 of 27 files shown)


kitchoi:
@rahulporuri Should I close this and just update the `setup.cfg` with a reference separately or is this easy enough to be reviewed?

rahulporuri:
It's easy enough to be reviewed but it'll just be one more commit before we possibly remove all of the code. I'm sorry for the confusion but it might just be better to discuss #327 and then update `setup.cfg` or any other solution.

kitchoi:
No problem. I did not spend a lot of time on actually flaking the thing: I spent more time understanding the issues surrounding whether `single_project` can be removed, which is probably worthwhile anyway.
Closing (I don't want to let PRs go stale, especially ones that touch a lot of files, it is easy to restore if needed).

PR #336, opened 2020-10-21
Add wx as being supported in etstool, add it back to CI, and test against wxPython v4.x 

### LINKED ISSUE #312: Test against wxPython v4.x instead of v3.x
At the moment, we seem to be testing against v 3.x of wxPython where as ETS (Pyface/Traitsui) recommends using wxPython 4.x. We should be able to use code from the `etstool.py` module in traitsui/pyface to be able to install the relevant version of wxPython correctly.

https://github.com/enthought/envisage/blob/ae477a4e4919681006c9e84bb4f9edb77b65bba8/etstool.py#L129.

aaronayres35:
fixes #312 

Note that previously, "wx" was listed in `available_toolkits`, but not in `supported_combinations` so trying to run `python etstool.py install --toolkit=wx` resulted in a failure.  Additionally, wx had been completely removed from CI altogether. (see #225)

This PR updates `etstool.py` so that wx is included in `supported_combinations` so that one can actually install and run tests with wx as the toolkit.  It also re-adds wx a as a toolkit to be run in CI.  Also, to address the original issue, it installs the relevant version of wx by following what is done in traitsui:
https://github.com/enthought/traitsui/blob/d112a694a4d681861a7ad3abb513cf7c83da9fe5/etstool.py#L246-L255

If we wish to continue simply not having wx as an available toolkit, we can close this PR.  If we do that we would want to open a separate PR removing "wx" references from `etstool.py`.  The only other occurrences of "wx" in envisage itself are here:
https://github.com/enthought/envisage/blob/215195ee5fb0285bc514ce40a573ea09aed69d11/examples/plugins/workbench/AcmeLab/acme/workbench/view/color_view.py#L58-L66
https://github.com/enthought/envisage/blob/215195ee5fb0285bc514ce40a573ea09aed69d11/examples/plugins/workbench/AcmeLabUsingEggs/src/acme.workbench/acme/workbench/view/color_view.py

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,6 +13,8 @@ addons:
     - libxcb-randr0
     - libxcb-render-util0
     - libxcb-xinerama0
+    # Wx dependencies
+    - libsdl2-2.0-0
 
 env:
   global:
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,9 +22,9 @@ env:
 
 matrix:
   include:
-    - env: RUNTIME=3.6 TOOLKITS="null pyqt5 pyside2"
+    - env: RUNTIME=3.6 TOOLKITS="null pyqt5 pyside2 wx"
     - os: osx
-      env: RUNTIME=3.6 TOOLKITS="null pyqt5 pyside2"
+      env: RUNTIME=3.6 TOOLKITS="null pyqt5 pyside2 wx"
 
 cache:
   directories:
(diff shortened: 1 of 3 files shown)


aaronayres35:
CI is now failing with the following error:
```
ImportError: libSDL2-2.0.so.0: cannot open shared object file: No such file or directory
```
Will look into this in the morning

rahulporuri:
Looks like you need to update the addon packages in travis for linux. See https://github.com/enthought/traitsui/blob/d112a694a4d681861a7ad3abb513cf7c83da9fe5/.travis.yml#L20-L21. Adding that should make CI happy.

rahulporuri (approved):
LGTM

PR #339, opened 2020-10-22
Use jinja templates for API documentation

### LINKED ISSUE #304: Consider using jinja templates for API documentation
ala https://github.com/enthought/traits-futures/pull/181. Ref traits-futures API docs https://docs.enthought.com/traits-futures/api/modules.html for comparison with envisage API docs https://docs.enthought.com/envisage/api.html or https://docs.enthought.com/envisage/api/modules.html

aaronayres35:
fixes #304 

This PR follows the lead of traits-futures, specifically PR https://github.com/enthought/traits-futures/pull/181

In doings so it adds templates and an `envisage.api.rst` file to `docs/source/spi`.  To avoid these files being deleted when building docs, the logic for removing any previously autogenerated API documentation has been modified.  The code was a bit cleaner before as we could just remove the full directory, and the logic in this PR may not be the best way to go about it.


etstool.py
```diff
@@ -435,14 +435,19 @@ def docs(edm, runtime, toolkit, environment):
...
+    parameters["templates_dir"] = "docs/source/api/templates/"
 
     # Remove any previously autogenerated API documentation.
-    if os.path.exists(docs_source_api):
-        rmtree(docs_source_api)
+    doc_api_files = os.listdir(docs_source_api)
+    permanent = ['envisage.api.rst', 'templates']
+    perviously_autogenerated = \
+        [file for file in doc_api_files if file not in permanent]
+    for file in perviously_autogenerated:
```
rahulporuri, line 445:
```suggestion
    previously_autogenerated = \
        [file for file in doc_api_files if file not in permanent]
    for file in previously_autogenerated:
```

rahulporuri (approved):
LGTM with one minor correction.

PR #340, opened 2020-10-22
Add documentation for envisage APIs

### LINKED ISSUE #302: Documentation for envisage.api is empty
Ref https://docs.enthought.com/envisage/api/envisage.html#module-envisage.api

aaronayres35:
fixes #302 

This PR updates the module docstrings of api modules so that their content is documented.

The PR is currently flagged as WIP as I am not sure the best way to organize the classes etc so that the documentation is more clear (specifically in the main `envisage.api` module).  ATM I have a `Interfaces` section and an `Exceptions` section, but I paused because I feel like this approach is not what we want long term.  I imagine we would want to break it up into things like `Plugins`, `ExtensionPoints` and `Services`, or maybe have sections for `Managers` or `Registry`s.  I am not sure, but will update this PR once this is decided on.


rahulporuri (changes requested):
I'm not sure exactly why but the built documentation is still empty.

aaronayres35:
> I'm not sure exactly why but the built documentation is still empty.

When I built it locally I was seeing it.  Very strange, I will investigate

aaronayres35:
> > I'm not sure exactly why but the built documentation is still empty.
> 
> When I built it locally I was seeing it. Very strange, I will investigate

I can't seem to reproduce.  @kitchoi if you have the time would you be able to try building the docs and lmk if the api modules are populated?

kitchoi:
Yep, I can try that.

kitchoi:
I can also see the populated content for the `api` modules.
This is what I did:
```
$ git checkout doc-302-document-apis
$ python etstool.py install --editable
$ python etstool.py docs
```
Then open `docs/build/index.html` for the main page, and navigate to the corresponding api modules...
e.g.
Screenshot 2020-10-23 at 15 54 26

envisage/api.py
```diff
@@ -8,6 +8,51 @@
...
+- :class:`~.IExtensionPoint`
+- :class:`~.IExtensionPointUser`
+- :class:`~.IExtensionProvider`
+- :class:`~.IExtensionRegistry`
+- :class:`~.IImportManager`
+- :class:`~.IPlugin`
+- :class:`~.IPluginActivator`
+- :class:`~.IPluginManager`
+- :class:`~.IServiceRegistry`
+
```
rahulporuri, line 27:
```suggestion

Application, plugin and related classes
------------------------------------------
```

rahulporuri (approved):
LGTM.

rahulporuri:
> I'm not sure exactly why but the built documentation is still empty.

Sorry about that. I always forget to install the package in editable mode.

PR #345, opened 2020-10-25
Fix Unregister service value error bug

### LINKED ISSUE #251: Unregistering service results in ValueError when application stops
I am getting a `ValueError` when stopping my application after unregistering a service provided by a `ServiceOffer` in another plugin. 

After some digging, it looks like the `CorePlugin._service_ids` contains a list of service IDs that were provided to the `'envisage.service_offers'` extension point. Then when the plugin stops, it attempts to unregister those services. However, if one of those services was unregistered with `Application.unregister_service(<id>)`, the id is not also removed from `CorePlugin._service_ids` and stopping the plugin results in the following:

```
  File "/Users/jcorson/.edm/envs/foo/lib/python3.6/site-packages/envisage/application.py", line 338, in stop
    self.plugin_manager.stop()
  File "/Users/jcorson/.edm/envs/foo/lib/python3.6/site-packages/envisage/composite_plugin_manager.py", line 167, in stop
    self.stop_plugin(plugin)
  File "/Users/jcorson/.edm/envs/foo/lib/python3.6/site-packages/envisage/composite_plugin_manager.py", line 177, in stop_plugin
    plugin.activator.stop_plugin(plugin)
  File "/Users/jcorson/.edm/envs/foo/lib/python3.6/site-packages/envisage/plugin_activator.py", line 49, in stop_plugin
    plugin.unregister_services()
  File "/Users/jcorson/.edm/envs/foo/lib/python3.6/site-packages/envisage/plugin.py", line 285, in unregister_services
    for service_id in service_ids:
  File "/Users/jcorson/.edm/envs/foo/lib/python3.6/site-packages/envisage/application.py", line 422, in unregister_service
    self.service_registry.unregister_service(service_id)
  File "/Users/jcorson/.edm/envs/foo/lib/python3.6/site-packages/envisage/service_registry.py", line 191, in unregister_service
    raise ValueError('no service with id <%d>' % service_id)
ValueError: no service with id <13>
```

aaronayres35:
fixes #251 

This PR first adds a failing regression test for the error described in the issue.  

To fix the issue, I added a listener in the `CorePlugin` class that listens to the `service_registry`'s `unregistered` event trait.  As mentioned in the issue, when `Application.unregister_service(<id>)` was called, the id was not also removed from `CorePlugin._service_ids`, so in the handler we simply remove the unregistered service id from `self._service_ids`.

### DIFF
--- a/envisage/plugin.py
+++ b/envisage/plugin.py
@@ -289,10 +289,10 @@ class Plugin(ExtensionProvider):
         service_ids.reverse()
 
         for service_id in service_ids:
-            # note the service may have already been individually unregistered
             try:
-                self.application.service_registry.get_service_from_id(service_id)  # noqa: E501
+                self.application.get_service_from_id(service_id)
             except ValueError:
+                # the service may have already been individually unregistered
                 pass
             else:
                 self.application.unregister_service(service_id)
--- a/envisage/plugin.py
+++ b/envisage/plugin.py
@@ -291,7 +291,7 @@ class Plugin(ExtensionProvider):
         for service_id in service_ids:
             # note the service may have already been individually unregistered
             try:
-                self.application.service_registry.get_service_from_id(service_id)
+                self.application.service_registry.get_service_from_id(service_id)  # noqa: E501
             except ValueError:
                 pass
             else:
--- a/envisage/plugin.py
+++ b/envisage/plugin.py
@@ -289,7 +289,12 @@ class Plugin(ExtensionProvider):
         service_ids.reverse()
 
         for service_id in service_ids:
-            self.application.unregister_service(service_id)
+            try:
+                self.application.service_registry.get_service_from_id(service_id)
+            except ValueError:
+                pass
+            else:
+                self.application.unregister_service(service_id)
 
         # Just in case the plugin is started again!
         self._service_ids = []
--- a/envisage/plugin.py
+++ b/envisage/plugin.py
@@ -289,6 +289,7 @@ class Plugin(ExtensionProvider):
         service_ids.reverse()
 
         for service_id in service_ids:
+            # note the service may have already been individually unregistered
             try:
                 self.application.service_registry.get_service_from_id(service_id)
             except ValueError:
--- a/envisage/tests/test_core_plugin.py
+++ b/envisage/tests/test_core_plugin.py
@@ -216,3 +216,24 @@ class CorePluginTestCase(unittest.TestCase):
 
         # Run it!
         return application.run()
+
+    def test_unregister_service(self):
+
+        class IJunk(Interface):
+            trash = Str()
+
+        class Junk(HasTraits):
+            trash = Str("garbage")
+        
+        some_junk = Junk()
+
+        application = TestApplication(
+            plugins=[CorePlugin()],
+        )
+        
+        application.start()
+
+        some_junk_id = application.register_service(IJunk, some_junk)
+        application.unregister_service(some_junk_id)
+
+        application.stop()
--- a/envisage/tests/test_core_plugin.py
+++ b/envisage/tests/test_core_plugin.py
@@ -192,7 +192,7 @@ class CorePluginTestCase(unittest.TestCase):
             service_offers = List(contributes_to=SERVICE_OFFERS)
 
             def _service_offers_default(self):
-            
+
                 a_service_offer = ServiceOffer(
                     protocol=IJunk,
                     factory=self._create_junk_service,
@@ -209,7 +209,7 @@ class CorePluginTestCase(unittest.TestCase):
             def _unregister_junk_service(self):
                 # only 1 service is registered so it has service_id of 1
                 self.application.unregister_service(1)
-        
+
         application = TestApplication(
             plugins=[CorePlugin(), PluginA()],
         )
@@ -224,13 +224,13 @@ class CorePluginTestCase(unittest.TestCase):
 
         class Junk(HasTraits):
             trash = Str("garbage")
-        
+
         some_junk = Junk()
 
         application = TestApplication(
             plugins=[CorePlugin()],
         )
-        
+
         application.start()
 
         some_junk_id = application.register_service(IJunk, some_junk)
--- a/envisage/tests/test_core_plugin.py
+++ b/envisage/tests/test_core_plugin.py
@@ -18,20 +18,7 @@ from pkg_resources import resource_filename
 # Enthought library imports.
 from envisage.api import Application, CorePlugin, Plugin
 from envisage.api import ServiceOffer
-from traits.api import (
-    HasTraits,
-    Interface,
-    List,
-    on_trait_change,
-    pop_exception_handler as pop_on_trait_change_handler,
-    push_exception_handler as push_on_trait_change_handler,
-    Str,
-)
-from traits.observation.api import (
-    pop_exception_handler as pop_observe_handler,
-    push_exception_handler as push_observe_handler,
-)
-
+from traits.api import HasTraits, Interface, List, on_trait_change, Str
 
 # This module's package.
 PKG = "envisage.tests"
@@ -45,14 +32,6 @@ class TestApplication(Application):
 
 class CorePluginTestCase(unittest.TestCase):
     """ Tests for the core plugin. """
-
-    def setUp(self):
-        push_on_trait_change_handler(reraise_exceptions=True)
-        push_observe_handler(reraise_exceptions=True)
-
-    def tearDown(self):
-        pop_observe_handler()
-        pop_on_trait_change_handler()
 
     def test_service_offers(self):
         """ service offers """
@@ -198,6 +177,10 @@ class CorePluginTestCase(unittest.TestCase):
 
     # regression test for enthought/envisage#251
     def test_unregister_service_offer(self):
+        """ Unregister a service that is contributed to the
+        "envisage.service_offers" extension point while the application is
+        running.
+        """
 
         class IJunk(Interface):
             trash = Str()
@@ -235,9 +218,12 @@ class CorePluginTestCase(unittest.TestCase):
         )
 
         # Run it!
-        return application.run()
+        application.run()
 
     def test_unregister_service(self):
+        """ Unregister a service which was registered on the application
+        directly, not through the CorePlugin's extension point. CorePlugin
+        should not do anything to interfere. """
 
         class IJunk(Interface):
             trash = Str()
--- a/envisage/tests/test_core_plugin.py
+++ b/envisage/tests/test_core_plugin.py
@@ -18,7 +18,19 @@ from pkg_resources import resource_filename
 # Enthought library imports.
 from envisage.api import Application, CorePlugin, Plugin
 from envisage.api import ServiceOffer
-from traits.api import HasTraits, Interface, List, on_trait_change, Str
+from traits.api import (
+    HasTraits,
+    Interface,
+    List,
+    on_trait_change,
+    pop_exception_handler as pop_on_trait_change_handler,
+    push_exception_handler as push_on_trait_change_handler,
+    Str,
+)
+from traits.observation.api import (
+    pop_exception_handler as pop_observe_handler,
+    push_exception_handler as push_observe_handler,
+)
 
 
 # This module's package.
@@ -33,6 +45,14 @@ class TestApplication(Application):
 
 class CorePluginTestCase(unittest.TestCase):
     """ Tests for the core plugin. """
+
+    def setUp(self):
+        push_on_trait_change_handler(reraise_exceptions=True)
+        push_observe_handler(reraise_exceptions=True)
+
+    def tearDown(self):
+        pop_observe_handler()
+        pop_on_trait_change_handler()
 
     def test_service_offers(self):
         """ service offers """
--- a/envisage/tests/test_core_plugin.py
+++ b/envisage/tests/test_core_plugin.py
@@ -16,9 +16,9 @@ import unittest
 from pkg_resources import resource_filename
 
 # Enthought library imports.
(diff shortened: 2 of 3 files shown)


aaronayres35:
My initial attempted solution is flawed as we dont actually want to lister for all unregistered services, but instead only those which were contributed as service offers to the `CorePlugin`'s `service_offers` extension point.  

Also note the comment here:
```
@on_trait_change("service_offers_items")
    def _service_offers_changed(self, event):
        """ React to new service offers being *added*.

        Note that we don't currently do anything if services are *removed* as
        we have no facility to let users of the service know that the offer
        has been retracted.

        """
        for service in event.added:
            self._register_service_offer(service)

        return
```

envisage/tests/test_core_plugin.py
```diff
@@ -187,3 +175,65 @@ def _preferences_default(self):
...
+                self.application.unregister_service(1)
+
+        application = TestApplication(
+            plugins=[CorePlugin(), PluginA()],
+        )
+
+        # Run it!
+        return application.run()
+
+    def test_unregister_service(self):
```
aaronayres35, line 220: This test never actually failed with the new fix, but while running I saw the following traceback:
```
Exception occurred in traits notification handler for event object: TraitChangeEvent(object=<envisage.service_registry.ServiceRegistry object at 0x7f993b991e60>, name='unregistered', old=<undefined>, new=1)
  File "/Users/aayres/.edm/envs/envisage-test-3.6-pyqt5/lib/python3.6/site-packages/traits/observation/_trait_event_notifier.py", line 122, in __call__
    self.dispatcher(handler, event)
  File "/Users/aayres/.edm/envs/envisage-test-3.6-pyqt5/lib/python3.6/site-packages/traits/observation/observe.py", line 26, in dispatch_same
    handler(event)
  File "/Users/aayres/Desktop/envisage/envisage/core_plugin.py", line 202, in unregistered_updated
    self._service_ids.remove(event.new)
  File "/Users/aayres/.edm/envs/envisage-test-3.6-pyqt5/lib/python3.6/site-packages/traits/trait_list_object.py", line 801, in remove
    super().remove(value)
  File "/Users/aayres/.edm/envs/envisage-test-3.6-pyqt5/lib/python3.6/site-packages/traits/trait_list_object.py", line 460, in remove
    super().remove(value)
ValueError: list.remove(x): x not in list
```
This test should probably be written so that "failures" like this do not go unnoticed!

kitchoi: Adding something like this in the test case's setup and teardown will allow the test to fail:
```
from traits.api import pop_exception_handler, push_exception_handler
...

    def setUp(self):
        push_exception_handler(reraise_exceptions=True)

    def tearDown(self):
        pop_exception_handler()

```

aaronayres35: hm I added this, but the test is still passing when I remove the fix.  I see:
```
$ python -m unittest test_core_plugin.py 
....Exception occurred in traits notification handler for event object: TraitChangeEvent(object=<envisage.service_registry.ServiceRegistry object at 0x7fb7fb052d00>, name='unregistered', old=<undefined>, new=1)
  File "/Users/aayres/.edm/envs/envisage-test-3.6-pyqt5/lib/python3.6/site-packages/traits/observation/_trait_event_notifier.py", line 122, in __call__
    self.dispatcher(handler, event)
  File "/Users/aayres/.edm/envs/envisage-test-3.6-pyqt5/lib/python3.6/site-packages/traits/observation/observe.py", line 26, in dispatch_same
    handler(event)
  File "/Users/aayres/Desktop/envisage/envisage/core_plugin.py", line 206, in unregistered_updated
    self._service_ids.remove(event.new)
  File "/Users/aayres/.edm/envs/envisage-test-3.6-pyqt5/lib/python3.6/site-packages/traits/trait_list_object.py", line 801, in remove
    super().remove(value)
  File "/Users/aayres/.edm/envs/envisage-test-3.6-pyqt5/lib/python3.6/site-packages/traits/trait_list_object.py", line 460, in remove
    super().remove(value)
ValueError: list.remove(x): x not in list
..
Ran 6 tests in 0.033s

OK
```

aaronayres35: ah it seems it is likely related to https://github.com/enthought/traits/issues/1323?

aaronayres35: > ah it seems it is likely related to [enthought/traits#1323](https://github.com/enthought/traits/issues/1323)?

Using both as suggested in the issue fixed the problem.  I also am realizing I used observe in `CorePlugin` for the fix, but in the new test I wrote I used `on_trait_change` without even realizing.  Should we always be using observe for any new code at this point? @kitchoi

kitchoi: > Should we always be using observe for any new code at this point?

Not sure, as this means the next release of envisage will drop support of Traits 6.0, but given envisage is higher up the stack, this might be okay. We should discuss and decide though. To avoid having this PR tied with an unresolved discussion, I'd probably use `on_trait_change` for now.

kitchoi:
The `_service_ids` is defined in `start`, but instead of having it reset in `stop`, we are removing values via a change handler. The asymmetry and complexity in the logic make me wonder if this is the right fix. If this is the right fix, it does not look _obviously_ right. @aaronayres35 Could you explain a bit more why this is done this way please?

kitchoi:
> The asymmetry and complexity in the logic make me wonder if this is the right fix. If this is the right fix, it does not look obviously right.

To elaborate... a more obviously correct fix is to implement `stop` that obviously reverses what `start` does.  Maybe we can have an `_unregister_service_offers` on the core plugin and reverse the action of registering service offers...  I have not explored if this could work, and maybe that is something you have tried already. But it would be good, for reference, to know if that more obviously correct alternative could fix the issue.

aaronayres35:
> The `_service_ids` is defined in `start`, but instead of having it reset in `stop`, we are removing values via a change handler. The asymmetry and complexity in the logic make me wonder if this is the right fix. If this is the right fix, it does not look _obviously_ right. @aaronayres35 Could you explain a bit more why this is done this way please?

So basically, what happened previously was that `_service_ids` was defined in start, by registering each service offer that was offered to the `service_offers` extension point.  And there is a comment saying:
```
# These services are unregistered by the default plugin activation
# strategy (due to the fact that we store the service ids in this
# specific trait!).
```
So, when `application.stop` was called it lead to `self.plugin_manager.stop()` being called which for each plugin called `plugin.activator.stop_plugin(plugin)`.  This then calls `plugin.stop()` and then `plugin.unregister_services()`.  Currently the `CorePlugin` does not override the base plugin `stop` method, so this does nothing.  `plugin.unregister_services()` is where the error occurs because if a service_offer is unregistered before this process has started it is not removed from the `CorePlugin`s `_service_ids`.

It would also be a valid approach to define a `stop` method on the `CorePlugin` class that tried to remove and service ids for services which were unregistered while the application was running from `_service_ids`.  I thought adding a listener would make more sense as in that case, the `_service_ids` is actually a kept up-to-date list of the service ids that are currently registered via the `service_offers` extension point.  


Note that if a service offer is not unregistered _while the application is running_ then they with out any of the changes made in this PR they will still be unregistered when `application.stop` is called via the ultimate call to ``plugin.unregister_services()`.  The problem only arises when unregistering a service before the application is stopped, as in that case when that happens, `service_ids` is not updated to match the current state of services that are registered.  In that sense I thought a listener was the best way to keep `_service_ids` always in sync.  It may be possible to replace the listener with a stop method on CorePlugin though and try to do the resynchronization then?  That seems a bit more confusing to me personally though

kitchoi:
Thank you @aaronayres35.

Looks like we are having more than one sources of truth here. Both the service registry on the application and the individual plugin are trying to manage the life-time of a service, so they could end up stepping on each other's toes, and causing a service to be unregistered twice (or more times). 

Eliminating multiple sources of truth would be a better fix. Otherwise, we are going to have to do more of these sychronization, e.g. not just `_service_ids` on the CorePlugin requires synchronization, any instances of Plugin that also maintains the service id list will also need to be synchronized.

kitchoi:
Another alternative approach to consider may be for plugins to anticipate that when it tries to unregister a service, it may have already been unregistered by others, and handle the exception there.

kitchoi:
> Eliminating multiple sources of truth would be a better fix. Otherwise, we are going to have to do more of these sychronization, e.g. not just _service_ids on the CorePlugin requires synchronization, any instances of Plugin that also maintains the service id list will also need to be synchronized.
> Another alternative approach to consider may be for plugins to anticipate that when it tries to unregister a service, it may have already been unregistered by others, and handle the exception there.

Ah, this would present a mental shift in understanding... The service ids on the plugins provide no indications as to whether the service is still used or not, they are mere references. The service registry is the one that has that information. Plugins are keeping these ids so that if they are independently deactivated, they can ask the service registry to unregister its service offers.  If this understanding holds, it would make sense for a plugin to expect and handle the exception when attempt to unregister a service fails because it is already gone.

aaronayres35:
> Ah, this would present a mental shift in understanding... The service ids on the plugins provide no indications as to whether the service is still used or not, they are mere references. The service registry is the one that has that information. Plugins are keeping these ids so that if they are independently deactivated, they can ask the service registry to unregister its service offers. If this understanding holds, it would make sense for a plugin to expect and handle the exception when attempt to unregister a service fails because it is already gone.

I see, maybe we could have a quick chat about this either before, after or during the standup today?  Just to make sure my understanding is on the same page.
Trying to follow this suggestion, in the `plugin.unregister_services` method I simply made the following change:
```
    def unregister_services(self):
        """ Unregister any service offered by the plugin. """

        # Unregister the services in the reverse order that we registered
        # them.
        service_ids = self._service_ids[:]
        service_ids.reverse()

        for service_id in service_ids:
            try:
                self.application.unregister_service(service_id)
            except ValueError:
                pass

        # Just in case the plugin is started again!
        self._service_ids = []

        return
```
and I removed my listener is `CorePlugin`.  With this change, all the tests pass.  Basically if we see a `ValueError` when unregistering a service as a plugin we basically just assume that that service was already unregistered (by the `service_registry`).  This is a bit ugly and there is probably a nicer way to do this, it was more just a sanity check of the different approach.

envisage/tests/test_core_plugin.py
```diff
@@ -34,11 +46,17 @@ class TestApplication(Application):
 class CorePluginTestCase(unittest.TestCase):
     """ Tests for the core plugin. """
 
+    def setUp(self):
+        push_on_trait_change_handler(reraise_exceptions=True)
+        push_observe_handler(reraise_exceptions=True)
+
+    def tearDown(self):
+        pop_observe_handler()
```
kitchoi, line 54: Nothing here uses `observe`, so we can remove the push and pop for the observe handleres.
In fact, is any of these `push_*` and `pop_*` required for reproducing the error if we remove the fix?

envisage/plugin.py
```diff
@@ -289,7 +289,13 @@ def unregister_services(self):
         service_ids.reverse()
 
         for service_id in service_ids:
-            self.application.unregister_service(service_id)
+            # note the service may have already been individually unregistered
+            try:
+                self.application.service_registry.get_service_from_id(service_id)  # noqa: E501
```
kitchoi, line 294: I suspect Plugin is not supposed to use `service_registry` on the `application` directly, and it would be more symmetric if we call `self.application.get_service_from_id` here instead before we call `self.application.unregister_service`. (I know at the end of the day they are the same after all the redirection. It just looks asymmetric and inconsistent...)

envisage/tests/test_core_plugin.py
```diff
@@ -187,3 +195,65 @@ def _preferences_default(self):
 
         # Make sure we can get one of the preferences.
         self.assertEqual("42", application.preferences.get("enthought.test.x"))
+
+    # regression test for enthought/envisage#251
+    def test_unregister_service_offer(self):
+
```
kitchoi, line 201: Could we add a docstring here (to be consistent)?

envisage/tests/test_core_plugin.py
```diff
@@ -187,3 +195,65 @@ def _preferences_default(self):
...
+            def _unregister_junk_service(self):
+                # only 1 service is registered so it has service_id of 1
+                self.application.unregister_service(1)
+
+        application = TestApplication(
+            plugins=[CorePlugin(), PluginA()],
+        )
+
+        # Run it!
+        return application.run()
```
kitchoi, line 238: Nit: The `return` is redundant here.

envisage/plugin.py
```diff
@@ -289,7 +289,13 @@ def unregister_services(self):
         service_ids.reverse()
 
         for service_id in service_ids:
-            self.application.unregister_service(service_id)
+            # note the service may have already been individually unregistered
```
kitchoi, line 292: Nit: This comment may be better placed in the `except ValueError` block. It directly explains why we are excusing the exception. A service can be unregistered by others and the plugin won't know about it. The ValueError indicates it is already unregistered, and the plugin is cool about it.

envisage/tests/test_core_plugin.py
```diff
@@ -187,3 +195,65 @@ def _preferences_default(self):
...
+
+    def test_unregister_service(self):
+
```
kitchoi, line 241: Same here, a docstring would be good, some hints to why this test is not duplicating the one above may also be helpful.

kitchoi (commented):
Some minor comments and questions.
(The test looks great, they are not junk 😂 .)

envisage/tests/test_core_plugin.py
aaronayres35: Ah, yes I believe they can all be removed now

kitchoi (approved):
LGTM

PR #347, opened 2020-10-26
[Draft] Reimplement extension registry using observe

kitchoi:
This PR demonstrates how the `ExtensionRegistry` can be rewritten using `observe` from Traits 6.1.

Instead of rewriting the `ExtensionRegistry`, I added an `ObservableExtensionRegistry` which reimplements the `ExtensionRegistry` interface using `observe`.

I shifted all of the tests on `ExtensionRegistry` to mixin classes and run the same tests again this new extension registry.  I also added more tests. These tests demonstrate one discrepancy in `ObservableExtensionRegistry`: a weakref is held for a nonmethod handler in ExtensionRegistry, but for ObservableExtensionRegistry, a strong reference will be held. However the weakref handling for nonmethod handler looks unnecessary in ExtensionRegistry, as such handling is immediately countered in the only usage I could find:
https://github.com/enthought/envisage/blob/8245931f7e3dfcb6513bfb7dc2ae9b10ca6f89a1/envisage/extension_point.py#L226-L229

In summary, I think the approach has worked well...

The original motivation was to find a solution to #291, and I tried to explore whether reimplementing the extension registry using observe and TraitList would provide a fix... 

The answer is No. :) Therefore I don't intend for this PR to be merged, I will close this after CI is green.

### DIFF
--- a/envisage/extension_registry.py
+++ b/envisage/extension_registry.py
@@ -11,14 +11,19 @@
 
 
 # Standard library imports.
+import functools
 import logging
 import types
 import weakref
 
 # Enthought library imports.
-from traits.api import Dict, HasTraits, provides
+from traits.api import (
+    ComparisonMode, Dict, HasTraits, Instance, List, provides, Str, Tuple,
+)
+from traits.observation.api import trait, observe
 
 # Local imports.
+from .extension_point import ExtensionPoint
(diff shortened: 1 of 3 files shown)


kitchoi:
> The answer is No. :) Therefore I don't intend for this PR to be merged, I will close this after CI is green.

Closing... :)

PR #348, opened 2020-10-26
Add apptools to source_dependencies in etstool.py

aaronayres35:
In #162 cron jobs were set up and the framework was added so that they could installed ets packages from source.  `apptools` was not included originally in the list of `source_dependencies`, but it should be.


rahulporuri (approved):
LGTM

PR #354, opened 2020-10-28
Support observe("name:items") for ExtensionPoint [Requires Traits 6.1]

### LINKED ISSUE #291: Support for observe in ExtensionPoint

kitchoi:
Closes #291

This PR allows `on_trait_change("name_items")` to be migrated to `observe("name:items")` where the `name` is defined as an `ExtensionPoint` trait.

After considering all the constraints and alternative approach (see #291), I have come to the conclusion that the support needs to happen by caching the TraitList returned for the ExtensionPoint and making it impossible to mutate the list directly.  Having a persisted list that can change is, strictly speaking, what `on_trait_change("x_items")` or `observe("x:items")` assumes.

This means the list of extensions has to be maintained not just on the extension registry, but on the object with the extension point as well.  This requires fair amount of efforts to keep these values in-sync. This PR still treats the extension registry as the single source of truth as to what extensions are available. The ExtensionPoint remains to be a proxy to query values in an extension registry.

Caution: This implementation requires Traits 6.1, as it imports `TraitList` from `traits.trait_list_object`.  We should decide if it is necessary to phase in Traits 6.1. Should it be decided that the next release of Envisage will still support Traits 6.0, some additional compatibility layer will be needed for this PR. Edited: ~but I don't think it would be much (We just need to roll the old ExtensionPoint definition for that, that's all.)~, or this PR can be considered for merging after next release.


envisage/plugin.py
```diff
@@ -327,7 +327,7 @@ def _anytrait_changed(self, trait_name, old, new):
             else:
                 added = new
                 removed = old
-                index = slice(0, max(len(old), len(new)))
+                index = slice(0, len(old))
```
kitchoi, line 330: This is a bug fix required for this PR. I should open a separate PR for this.

envisage/tests/test_extension_point.py
```diff
@@ -120,21 +122,56 @@ def test_mutate_extension_point_no_effect(self):
...
+        with self.assertWarns(RuntimeWarning):
+            f.x.sort()
 
         # then
         # The registry is not changed, and the extension point is still the
         # same as before
-        self.assertEqual(registry.get_extensions("my.ep"), [1, 2, 3])
-        self.assertEqual(f.x, [1, 2, 3])
+        self.assertEqual(registry.get_extensions("my.ep"), [1, 2, 3, 0, 5])
+        self.assertEqual(f.x.copy(), [1, 2, 3, 0, 5])
```
kitchoi, line 174: The change to `f.x.copy()` makes the test more strict and meaningful because it falls back to getting the actual values on the list, otherwise the `__eq__` logic is used for this comparison and we would just be comparing the values on the registry.

envisage/tests/test_extension_point_changed.py
```diff
@@ -188,37 +262,29 @@ def test_assign_empty_list(self):
...
-        # Make sure we pick up the correct contribution via the plugin.
-        extensions = a.x[:]
-        extensions.sort()
-
-        self.assertEqual(3, len(extensions))
-        self.assertEqual([98, 99, 100], extensions)
-
-        # We shouldn't get a trait event here because we haven't accessed the
-        # extension point yet!
-        self.assertEqual(None, listener.obj)
```
kitchoi, line 221: This test is removed because it is testing the behaviour noted in the "fixme" comment:
```
# fixme: If the extension point has not been accessed then the
# provider extension registry can't work out what has changed, so it
# won't fire a changed event.
``` 

The "fixme" is now fixed.

envisage/extension_point.py
```diff
@@ -200,23 +200,41 @@ def connect(self, obj, trait_name):
...
+                # For on_trait_change('name_items')
+                obj.trait_property_changed(
+                    trait_name + "_items", Undefined, event
+                )
 
-            return
+            else:
+                # The entire list has changed. We reset the cache and fire a
+                # normal trait changed event.
+                self._update_cache(obj, trait_name)
```
kitchoi, line 230: This causes the listener to hold a reference to the ExtensionPoint trait type instance. On its own, this is not a big deal, but then the trait type also tries to hold a reference to the listener to counter the weakref held in the extension registry (see the bottom of this `connect` function), so that results in a cycle reference.

envisage/extension_point.py
```diff
@@ -264,3 +282,278 @@ def _get_extension_registry(self, obj):
...
+        trait_name : str
+            The name of the trait for which ExtensionPoint is defined.
+        """
+        super().__init__(iterable)
+
+        # Flag to control access for mutating the list. Only internal
+        # code can mutate the list. See _sync_values
+        self._internal_use = False
+
+        self._object = object
```
kitchoi, line 369: I was wondering if this has to be a weakref... and I am afraid it does to avoid cycle reference, and I will need to implement the pickling logic for the weakref.

kitchoi (commented):
I will to check for cycle references here, marking as WIP for a bit.

envisage/extension_point.py
kitchoi: I am going to remove this: It only serves to support out-of-sync scenario when the listeners are not connected. This is probably not a scenario we need to worry about (?), I will note this as a limitation as the effort and maintenance burden to keep these back references are too much to be worth it.

kitchoi:
> I will to check for cycle references here, marking as WIP for a bit.

Done.

envisage/extension_point.py
```diff
@@ -200,23 +200,41 @@ def connect(self, obj, trait_name):
...
-
-            # If an index was specified then we fire an '_items' changed event.
+            """ Listener called when an extension point is changed.
+
+            Parameters
+            ----------
+            extension_registry : IExtensionRegistry
+                Registry that maintains the extensions.
+            event : ExtensionPointChangedEvent
+                Event for created for the change.
```
rahulporuri, line 210:
```suggestion
                Event created for the change.
```

envisage/extension_point.py
```diff
@@ -200,23 +200,41 @@ def connect(self, obj, trait_name):
...
+                If the event.index is None, this means the entire extensions
```
rahulporuri, line 211:
```suggestion
                If the event.index is None, this means the entire extensions list(?)
```

envisage/plugin.py
rahulporuri: I'm not sure what the bug is here. Can you elaborate?

kitchoi: oops, I failed to open that separate PR. I should do that and it would be easier to see it there :)

kitchoi: Done: https://github.com/enthought/envisage/pull/357

envisage/extension_point.py
kitchoi: Done (on the next line).

kitchoi:
From offline discussion: The current plan for the next release of Envisage (current version is 4.9.2) is that it will still support Traits 6.0. With that this PR should remain on hold until after the next release.

Review is still very welcome though.

envisage/extension_point.py
```diff
@@ -264,3 +282,249 @@ def _get_extension_registry(self, obj):
...
+        """ Reimplemented TraitList.__delitem__ """
+
+        # This is used by internal code
+
+        if not self._internal_use:
+            warnings.warn(
+                "Extension point cannot be mutated directly.",
+                RuntimeWarning,
+                stacklevel=2,
+            )
```
rahulporuri, line 363: is it overkill to refactor all of the warnings into a common `_raise_warnings` method or something of that sort?

rahulporuri (commented):
This LGTM but I would like to test this against internal apps which use envisage to confirm further.

Also, this is a substantial change and I think it might be good to put together a design document accompanying the changes. Do you think that's worth the time?

envisage/extension_point.py
kitchoi: Not an overkill at all... I should do that.

>  I think it might be good to put together a design document accompanying the changes.

I'd thought that extensive class docstring on `_ExtensionPointValue` was good enough to explain the rationale, the benefit compared to having a separate design document is that it is close to the code, and therefore it can describe concrete details that can be maintained together.  A design document tends to stay high-level and more abstract. Is there any high-level rationale I missed there and does not sit well in the class or module docstring?

rahulporuri: that makes sense too. i think class and module docstrings should be enough.

envisage/tests/test_extension_point.py
```diff
@@ -253,6 +328,80 @@ class Foo(TestBase):
...
+        f = ClassWithExtensionPoint(extension_registry=self.registry)
+        ExtensionPoint.connect_extension_point_traits(f)
+        self.registry.set_extensions("my.ep", [1, 2, 3])
+        self.assertEqual(f.x, [1, 2, 3])
+
+        # then
+        for protocol in range(pickle.HIGHEST_PROTOCOL + 1):
+            serialized = pickle.dumps(f.x, protocol=protocol)
+            deserialized = pickle.loads(serialized)
+            self.assertEqual(deserialized, [1, 2, 3])
```
kitchoi, line 403: When this test was run on Python 3.6 (the one built from `etstool.py`), the test passes.
When I try to run it on Python 3.8 and Python 3.9 after pip-installing envisage from source, this test fails:
```
======================================================================
FAIL: test_object_pickability (envisage.tests.test_extension_point.ExtensionPointTestCase)
  File "/Users/kchoi/Work/ETS/envisage/envisage/tests/test_extension_point.py", line 403, in test_object_pickability
    self.assertEqual(deserialized, [1, 2, 3])
AssertionError: [] != [1, 2, 3]
```

kitchoi: Same failure occurs to a Python 3.7 venv, when the test is run, this RuntimeWarning appears:
```
test_object_pickability (envisage.tests.test_extension_point.ExtensionPointTestCase) ... /Users/kchoi/Work/ETS/envisage/envisage/tests/test_extension_point.py:402: RuntimeWarning: Extension point cannot be mutated directly.
  deserialized = pickle.loads(serialized)
FAIL
...
======================================================================
FAIL: test_object_pickability (envisage.tests.test_extension_point.ExtensionPointTestCase)
  File "/Users/kchoi/Work/ETS/py37-venv/envisage-env/lib/python3.7/site-packages/envisage/tests/test_extension_point.py", line 403, in test_object_pickability
    self.assertEqual(deserialized, [1, 2, 3])
AssertionError: [] != [1, 2, 3]
```

kitchoi:
Refactored warning code and merged master in here. However the pickability on Python 3.7+ needs a bit more investigation.

envisage/tests/test_extension_point.py
kitchoi: Fixed.

kitchoi:
Fixed un/pickling issue. Note that CI is not run against Python 3.7+. I ran the test locally with a venv from Python 3.9 and verified they pass.

envisage/extension_point.py
```diff
@@ -192,21 +195,41 @@ def connect(self, obj, trait_name):
...
-                name = trait_name
-                old = event.removed
-                new = event.added
+                # The entire list has changed. We reset the cache and fire a
+                # normal trait changed event.
+                _update_cache(obj, trait_name)
 
-            obj.trait_property_changed(name, old, new)
+        # In case the cache was created first and the registry is then mutated
+        # before this ``connect``` is called, the internal cache would be in
```
rahulporuri, line 228:
```suggestion
        # before this ``connect`` is called, the internal cache would be in
```

rahulporuri (approved):
Changes LGTM. Spending a bit of time testing these changes with internal projects.

rahulporuri:
Tested against an internal application and things worked as expected. Still LGTM. I'll maybe spend a bit more time testing against one more internal application.

rahulporuri:
> Caution: This implementation requires Traits 6.1, as it imports TraitList from traits.trait_list_object. We should decide if it is necessary to phase in Traits 6.1. Should it be decided that the next release of Envisage will still support Traits 6.0, some additional compatibility layer will be needed for this PR. Edited: but I don't think it would be much (We just need to roll the old ExtensionPoint definition for that, that's all.), or this PR can be considered for merging after next release.

I think the next release of envisage will depend on traits >= 6.2.0 so we don't need to worry about adding any additional compatibility layer.

kitchoi:
> I think the next release of envisage will depend on traits >= 6.2.0 so we don't need to worry about adding any additional compatibility layer.

I will update the `setup.py` to require `traits>=6.1` so that the bump is tied to the change that requires it.

kitchoi:
> I will update the setup.py to require traits>=6.1 so that the bump is tied to the change that requires it.

Done. I believe the `setup.py` the only place that requires the bump, and I tested installation locally with `pip` successfully (CI does not have a job that uses `pip`). I did not change `etstool.py` as the EDM tool is already pulling in the latest released version of traits, and if it did not, the test suite will fail so we will know about it.

kitchoi:
Confirmed offline that this can be merged... merging...

PR #359, opened 2020-11-04
[WIP] Contributing a few examples to etsdemo

### LINKED ISSUE #320: Consider contributing a few examples to etsdemo.
For the 5.0.0 release, it might be good to contribute a few of the example envisage applications to the etsdemo application. The following examples work as expected and make for a good start : 

- hello world.
- MOTD.
- tasks attractors.

aaronayres35:
closes #320 

In this PR I currently attempt to contribute a few examples to etsdemo.  To do so, I added the entry point to `setup.py`, defined a function in the package to return info about the files, and copied the examples over to sit inside the package. 
Additionally, I removed the logging related code from the examples that were copied into the package.

I copied not moved the examples because I believe the documentation has links to them currently.  

Currently, the examples do not run. As is mentioned on the issue, they assume a specific Python Path when they are run which causes problems.   Modules seem to get renamed inside the file tree in the demo (e.g. `acme` is shown as `Acme`, underscores turn to spaces).  When clicking run with the `run.py` file selected for the `MOTD` example, the following traceback occurs:
```
  File "/Users/aayres/Desktop/traitsui/ets-demo/etsdemo/app.py", line 472, in run_code
    exec(self.source, locals, locals)
  File "<string>", line 5, in <module>
ModuleNotFoundError: No module named 'acme'
```
I'm unsure how to best resolve this problem. @kitchoi any suggestions would be appreciated.  My only idea currently is to define the relevant classes, etc. in just one file (for MOTD this isnt _too_ noisy), but this is not the right solution.

 However, the current `hello_world.py` example seems like it should run fine (no errors occur currently), but no output is printed.  It only imports from `traits.api` and `envisage.api`.  I'm unsure what the problem is here.


kitchoi:
> I copied not moved the examples because I believe the documentation has links to them currently.

I'd very much prefer not duplicating the examples because of the documentation links. We should try to update the documentation source so that we can move the examples instead of duplicating them.

There are two possible solutions to solving this issue:
(1) Temporarily modify `sys.path` when the application runs (see #377)
(2) Since the examples are moved into `envisage`, we can make use of namespace packages. e.g. we will have to change `from acme import ...` to `from envisage.examples.MOTD.acme import ...` instead. But places where package dependencies are evaluated at runtime, e.g. `ExtensionPoint(List(Instance("acme.motd.api.IMessage")))`, those will need to be changed as well, e.g. to `ExtensionPoint(List(Instance("envisage.examples.MOTD.acme.motd.api.IMessage")))`. Therefore I prefer option (1).

kitchoi:
> I'd very much prefer not duplicating the examples because of the documentation links. We should try to update the documentation source so that we can move the examples instead of duplicating them.

See #379

aaronayres35:
superseded by #380

PR #362, opened 2020-11-11
Import from pyface.qt 

### LINKED ISSUE #360: Don't import directly from PyQt
There are some places where we are directly importing from `PyQt4`.  We should avoid this because the code will not work if `PySide2` or `PyQt5` are being used.  Instead we should import via `pyface.qt`.

So for example:
https://github.com/enthought/envisage/blob/215195ee5fb0285bc514ce40a573ea09aed69d11/examples/plugins/workbench/AcmeLab/acme/workbench/view/color_view.py#L71

Should instead do:
```
from pyface.qt import QtGui
```

There are a few other places where this needs to be fixed.

anshsrtv:
### Fixes #360 

The import statements have been updated.


rahulporuri (approved):
LGTM. Thanks for the contribution @anshsrtv !

PR #377, opened 2020-11-30
Make example run from any directory

kitchoi:
This approach may provide for solving #320 and resolves the import issues seen in #359.

This PR changes the main scripts in the examples such that they can be run from another directory while still be able to import the local packages. The (dirty) trick is to temporarily insert the directory path where the script is to `sys.path` (which is how it works when one runs the script from that directory).


rahulporuri (commented):
Looks like we're missing `__init__` files in `envisage.examples` and `envisage.examples.tests`. Apart from that the examples ran as expected.

Did you also test if these work as expected when run via etsdemo?

kitchoi:
> Looks like we're missing __init__ files in envisage.examples and envisage.examples.tests. Apart from that the examples ran as expected.

Thank you!  Will be adding them back.

> Did you also test if these work as expected when run via etsdemo?

Yes, and the examples can be run from there. However since this PR does not move the files into `envisage` and does not add an entry point, it does not close #320 outright. I tested with the etsdemo using a local script like this:
```
import os
from etsdemo.main import main
main(
    [
        {
            "version": 1,
            "name": "Envisage Examples",
            "root": os.path.dirname(__file__),
        },
    ]
)
```
(where the script sits in the examples folder).

rahulporuri (approved):
LGTM. Do you have time to also work on a PR contributing the examples to etsdemo? If it can be done within a couple of days, i think it's worth holding the 5.0.0 release of envisage.

kitchoi:
> Do you have time to also work on a PR contributing the examples to etsdemo? If it can be done within a couple of days, i think it's worth holding the 5.0.0 release of envisage.

Yes it can be done in a couple of days. I can give this a try. Sorry for visiting the issue this late.

PR #380, opened 2020-12-01
Contribute examples to etsdemo

### LINKED ISSUE #320: Consider contributing a few examples to etsdemo.

aaronayres35:
closes #320 

replaces #359 

This PR takes the steps described in this comment: https://github.com/enthought/envisage/issues/320#issuecomment-736513808

It may be useful to review commit by commit as the commits generally follow each point in that comment.  

Also, I tried testing the updated documentation but links result in 404 errors as the github links will not exist until after this PR is merged.  

Further, at the moment, everything seems to run fine with `etsdemo` (no errors / everything looks correct), but when you run the examples (e.g. `hello_world.py`) and navigate to the output tab in the demo application, nothing is printed.  When running the example from the command line, "The hello.world application says Hola World!" is printed.  

(NOTE: the above happens when install etsdemo with pyside2.  when trying to install with pyqt5, the application fails to launch - which seems related to https://github.com/enthought/pyface/issues/418, and with wx, the application crashes once you navigate to the example file you want to run - e.g. `hello_world.py`)


kitchoi:
> but when you run the examples (e.g. hello_world.py) and navigate to the output tab in the demo application, nothing is printed. When running the example from the command line, "The hello.world application says Hola World!" is printed.

I have witnessed this yesterday. Upon further investigation, it is because etsdemo has invented a triple-under `___main___` (instead of `__main__`).
https://github.com/enthought/traitsui/blob/f7da6909ad1cecf457f61df5970c9281575ec61f/ets-demo/etsdemo/app.py#L468
Note that the fix is not just about changing `___main___` to `__main__`. That function is serving two purposes...

setup.py
```diff
@@ -317,6 +320,7 @@ def get_long_description():
         packages=find_packages(),
         package_data={
             "": ["images/*", "*.ini"],
+            "envisage": ["examples/*"],
```
kitchoi, line 323: Is this recursive? Could you check if the 'preferences.ini' in `plugins/tasks/attractors` (quite a few folders deep) file is included please?

Given `examples` is a package, you might want to split this out to make it more explicit what is data (vs packages, which get picked up by `find_packages` on line 320), e.g.:
```
"envisage.examples": [
    "Hello_World/*",
]
```
On the other hand, folders like "Hello_World" are now indistinguishable compared to packages such as `envisage.examples.tests`. It might be clearer to put things one level down, e.g. `envisage/examples/demo`.
Then you can do `"envisage.examples": ["demo/*", "demo/*/*", ...]`

aaronayres35: I went with the approach you described, but there are many nested folders so I ended up with:
```
"envisage.examples": [
                "demo/*",
                "demo/*/*",
                "demo/*/*/*",
                "demo/*/*/*/*",
                "demo/*/*/*/*/*",
                "demo/*/*/*/*/*/*",
            ],
```
in order to include everything

kitchoi: Like a little christmas tree 🎄 . I don't know of a better solution apart from doing an `os.path.walk(... topdown=False)` once to determine how deep the file structure is.

kitchoi: The christmas tree is probably good enough for now.

docs/source/conf.py
```diff
@@ -233,6 +233,6 @@
 
 extlinks = {
     'github-demo': (
-        'https://github.com/enthought/envisage/tree/master/examples/%s',   # noqa: E501
+        'https://github.com/enthought/envisage/tree/master/envisage/examples/%s',   # noqa: E501
```
kitchoi, line 236:
```suggestion
        'https://github.com/enthought/envisage/tree/master/envisage/examples/demo/%s',   # noqa: E501
```

envisage/_entry_point.py
```diff
@@ -0,0 +1,36 @@
...
+        Currently this is a placeholder.
+    Returns
+    -------
+    response : dict
+    """
+    return dict(
+        version=1,
+        name="Envisage Examples",
+        root=(
+            pkg_resources.resource_filename("envisage", "examples/demo")
```
kitchoi, line 34: Given the package data is now defined under `envisage.examples`, it would be more correct to say `pkg_resources.resource_filename("envisage.examples", "demo")`. Though, this would not fail, I think, as `pkg_resources` is not going to complain.

Could you add a test for this little function please?

(I think in the future we will migrate this over to importlib, but that's probably best dealt with separately with other current uses of `pkg_resources`).

kitchoi (commented):
Mostly LGTM. Thank you!
One suggested change for the doc link and a small request for a test.

envisage/_entry_point.py
aaronayres35: ah I remember now making similar changes in traits: https://github.com/enthought/traits/pull/1348

I will add a similar test to what was added in that PR, and I think I will also move and rename this `_entry_point.py` file so that its purpose is more clear.

kitchoi (approved):
LGTM. 

The `examples/GUI_Application` is also good enough to be exposed (confirmed with @rahulporuri). That is a very simple example, and I have tested it to work with etsdemo after the folder is moved to `envisage/examples/demo` too. We can simply move the folder in this PR or we could also do that in a separate PR.

I also opened #381 to consider removing the other examples that still live at the top-level, which are more visible than the ones moved into `envisage/examples`.

Upon merging of this PR, the live documentation on `docs.enthought.com/envisage` will have broken links. I opened #382 to track the broken links separately. But I expect the issue to be fixed along with the upcoming release so that is no extra work needed. The issue is open merely for tracking purposes, e.g. in case the release is delayed, we might want to rebuild the documentation earlier than the release.

setup.py
```diff
@@ -317,6 +320,14 @@ def get_long_description():
         packages=find_packages(),
         package_data={
             "": ["images/*", "*.ini"],
+            "envisage.examples": [
+                "demo/*",
+                "demo/*/*",
+                "demo/*/*/*",
+                "demo/*/*/*/*",
+                "demo/*/*/*/*/*",
+                "demo/*/*/*/*/*/*",
```
jwiggins, line 329: Wait, really? Is there no concise way to do this in a single string?

kitchoi: I couldn't find one. If there is a way, I'd love to know!

kitchoi:
This move is making the old examples at the top level more prominent than the relevant and good ones. So if this PR goes through, I would want to make sure we have notes in `examples/` (at the top level) to refer people to `envisage/examples` and make the old ones less obvious.  I find the in-between state worse than not exposing the examples in the demo app at all. So if we could not follow this through, I'd prefer we abandon this move.

kitchoi:
The quick and dirty way to get through this is perhaps:
Move the ones that are still in `examples/` (top-level) into another nested folder, name that folder something less attractive, and add a README in `examples/` to refer readers to `envisage/examples` instead. This way anyone browsing the repo would discover the ones in `envisage/examples` more easily.

kitchoi (approved):
Thank you!

PR #390, opened 2020-12-21
Change default pickle protocol to more recent version for python >= 3.4

aaronayres35:
This PR simply change the default value for `layout_save_protocol` to 4 as protocol version 4 is safe for python >= 3.4.  It also removes a test that was testing that protocol 2 was used by default, which we no longer want.

This was missed when addressing #313


kitchoi (approved):
LGTM. Could you please update the PR title to be more specific about bumping the pickle protocol? Changing default pickle protocol is not insignificant (though I so far don't see any harm).

PR #400, opened 2021-04-16
Replace "depends_on" in "Property" traits with "observe"

rahulporuri:
This PR replaces the `depends_on` in `Property` traits with `observe` instead. I tested that the attractors example still works as expected.


aaronayres35 (approved):
Changes LGTM

I tested the attractors example as well and generally things seemed okay.

However, if I change the `initial_point` of any of the models I get an error in the traceback.  I checked and the same thing was occurring on master, so it is through no fault of this PR.

:shipit:

rahulporuri:
created https://github.com/enthought/envisage/issues/404 for the attractors example issue

PR #410, opened 2021-05-12
Explicitly require traits 6.2

aaronayres35:
After the recent traitsUI fiasco, this was on my mind.

After #400 we now depend on traits 6.2+


rahulporuri (approved):
LGTM

PR #413, opened 2021-05-14
Release version 6.0.0 of envisage

rahulporuri:
This PR updates the version information to 6.0.0 and flips the `IS_RELEASED` flag. Once this commit is merged into the `maint/6.0` maintenance branch, the merged commit will be tagged.

Note that this PR is against the `maint/6.0` branch - meant for future bug-fix releases i.e. 6.0.x.


aaronayres35 (approved):
LGTM

~I haven't done any testing locally though~ I created an sdist, installed it in a fresh environment with `pip install envisage-6.0.0.tar.gz[test,ipython,ui]` and ran the test suite.  All tests passed (although 4 were skipped)

Curiously, the tests are skipped with the message "skipped 'Test requires a non-null GUI backend'", however even after running `pip install pyqt5` or `edm install pyqt5` they are still skipped.  This probably warrants an issue if one doesn't already exist, but shouldn't hold up the release.

rahulporuri:
> I haven't done any testing locally though

I've tested this locally. I ran the testsuite and there is one test failure out of the box - and that's because of a missing optional dependency (`ipykernel`) which isn't handled properly by the testsuite.

Apart from that, I have tested this with an internal envisage application and the testsuite of the internal app passes without any issues. The application itself also works as expected.

rahulporuri:
> Curiously, the tests are skipped with the message "skipped 'Test requires a non-null GUI backend'", however even after running pip install pyqt5 or edm install pyqt5 they are still skipped

`ETS_TOOLKIT` needs to be set appropriately for these tests to pass. Instead of checking for the presence of a toolkit, the tests are skipped depending on the presence of the `ETS_TOOLKIT` env var.

PR #422, opened 2021-06-11
Revert PR #354

rahulporuri:
This PR reverts PR #354 from the `master` branch. This PR seems to be the cause of the issue #417 which is why we are reverting it, until we can investigate the issue and find a solution.

This PR also undoes a change to the plugin test case made in PR #401. Because we cannot use `observe` to listen to changes in an extension point anymore, the conversion from `_*_changed` to `observe` needed to be undone.

This PR replaces #418 which only tries to revert PR #354 on the maintenance branch. We prefer this PR over #418 because we don't want the `master` branch to be in a broken state.

### DIFF
--- a/envisage/extension_point.py
+++ b/envisage/extension_point.py
@@ -11,14 +11,11 @@
 
 
 # Standard library imports.
-from functools import wraps
 import inspect
-import warnings
 import weakref
 
 # Enthought library imports.
 from traits.api import List, TraitType, Undefined, provides
-from traits.trait_list_object import TraitList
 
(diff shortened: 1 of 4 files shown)


rahulporuri:
@mdickinson i opened this PR given our earlier discussion.

aaronayres35 (approved):
LGTM

~although I haven't yet tested #421 with these changes yet~ EDIT: done, the test passes with these changes 🎉

mdickinson (approved):
LGTM; thank you.

PR #427, opened 2021-06-17
Backport PRs #421, #422 and #423

rahulporuri:
This PR backports the following PRs back to the `maint/6.0` branch in preparation for the 6.0.1 release - #421 , #422 and #423 . #422 reverted the original PR that is the cause of #417 , #421 adds a regression test for #417 and #423 ensures that the testsuite passes cleanly with minimal dependencies.

This PR also updates the changelog in preparation for the upcoming release.  Once this PR is merged, we will be in a place to make the 6.0.1 release.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -2,6 +2,28 @@
  Envisage CHANGELOG
 ====================
 
+Version 6.0.1
+=============
+
+Released: 2021-06-18
+
+This bugfix release fixes the issue where Extension Point resolution was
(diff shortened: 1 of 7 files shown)


aaronayres35 (approved):
LGTM

Looks like we settled on a bug fix not minor release.  That works for me

rahulporuri:
> Looks like we settled on a bug fix not minor release

Yup. We (@mdickinson ) settled on that because we're fixing a broken feature (the fix here is to effectively remove the feature).

PR #431, opened 2021-06-23
Revert recent changes to flake8 command in etstool.py

aaronayres35:
Follow up to #426, see comment: https://github.com/enthought/envisage/pull/426#pullrequestreview-690385930

These changes should no longer be needed as we don't mess with the edm root directory anymore on the gh action workflows.


mdickinson (approved):
LGTM. Thank you!

PR #446, opened 2021-08-09
 Default branch has been renamed from "master" to "main"

rahulporuri:
This PR replaces the use of `master` in the repo with `main` as the default branch name as been updated.

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -233,6 +233,6 @@ intersphinx_mapping = {
 
 extlinks = {
     'github-demo': (
-        'https://github.com/enthought/envisage/tree/master/envisage/examples/demo/%s',   # noqa: E501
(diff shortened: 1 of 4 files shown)


docs/source/conf.py
```diff
@@ -233,6 +233,6 @@
 
 extlinks = {
     'github-demo': (
-        'https://github.com/enthought/envisage/tree/master/envisage/examples/demo/%s',   # noqa: E501
+        'https://github.com/enthought/envisage/tree/main/envisage/examples/demo/%s',   # noqa: E501
```
rahulporuri, line 236: same comment as pyface/traitsui - we shouldn't be linking to examples on the `main` branch, we should be linking to specific versions!

mdickinson (approved):
LGTM

rahulporuri:
Update - same as with traitsui and chaco, github is automatically handling the redirect from `master` to `main` for the demo links in the documentation.

PR #465, opened 2022-05-14
Simplify ImportManager by using importlib

corranwebster:
Simple drive-by modernization as I was reading code.  Should have no effective change in behaviour.

I can't remove the other __import__ usage easily however as it is doing a `from ... import ...`-style import.

### DIFF
--- a/envisage/import_manager.py
+++ b/envisage/import_manager.py
@@ -9,6 +9,7 @@
 # Thanks for using Enthought open source!
 """ The default import manager implementation. """
 
+import importlib
 
 # Enthought library imports.
 from traits.api import HasTraits, provides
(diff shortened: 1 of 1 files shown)


mdickinson (approved):
LGTM. Do you know why we need a full `eval` there rather than just a `getattr`?

corranwebster:
> Do you know why we need a full `eval` there rather than just a `getattr`?

No idea.  If I had to guess it would be a hack to allow dotted imports like `foo.bar:Bar.baz`, but it would not shock me if it was used for more general things.

Obviously didn't want to touch it since I didn't know the scope of usage, but I'd be happy to see it go if it were safe.

PR #604, opened 2026-06-24
ci: pin third-party GitHub Actions to commit SHAs

mdickinson:
## Summary

Hardens CI supply-chain security by pinning all external GitHub Actions to immutable commit SHAs instead of mutable major-version tags, and adds a Dependabot cooldown.

### Why

A `uses: owner/repo@v7` reference resolves through a **mutable** git tag. If an attacker compromises an action's repository — or just gains the ability to move a tag — they can retarget `v7` at malicious code, and every workflow pinned to that tag runs it on the next CI run, with access to that job's secrets. This is exactly what happened in the **March 2025 `tj-actions/changed-files` compromise**, where version tags were retroactively moved to a commit that exfiltrated CI secrets. Full-SHA pins are immutable, so each run is reproducible and immune to tag-move attacks. This follows [GitHub's security-hardening guidance](https://docs.github.com/en/actions/security-for-github-actions/security-guides/security-hardening-for-github-actions#using-third-party-actions).

### Changes

- Pin all external actions to commit SHAs with a trailing version comment:

  | Action | SHA | Version |
  |---|---|---|
  | `actions/checkout` | `9c091bb` | v7.0.0 |
  | `actions/setup-python` | `ece7cb0` | v6.3.0 |
  | `actions/cache` | `27d5ce7` | v5.0.5 |
  | `actions/upload-artifact` | `043fb46` | v7.0.1 |
  | `enthought/setup-edm-action` | `15aaa25` | v4.1 |

- Add a **7-day Dependabot cooldown** (`cooldown: default-days: 7`) so action updates aren't picked up until they've been published for a week — reducing exposure to a compromised release that gets yanked shortly after publication.

The local `./.github/actions/install-qt-support` action is in-repo and needs no pinning. Dependabot continues to maintain the SHA pins, bumping both the SHA and the version comment.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,3 +4,5 @@ updates:
     directory: "/"
     schedule:
       interval: "weekly"
+    cooldown:
+      default-days: 7
--- a/.github/workflows/check-style-strict.yml
+++ b/.github/workflows/check-style-strict.yml
@@ -7,8 +7,8 @@ jobs:
     runs-on: 'ubuntu-latest'
 
     steps:
-    - uses: actions/checkout@v7
-    - uses: actions/setup-python@v6
+    - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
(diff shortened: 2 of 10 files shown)


mdickinson:
Trivial PR; self-merging.

PR #599, opened 2026-02-24
Remove EggPluginManager, EggBasketPluginManager, PackagePluginManager

mdickinson:
## Summary

Remove the deprecated `EggPluginManager`, `EggBasketPluginManager` and `PackagePluginManager` classes for the Envisage 8.0 release. These were deprecated in #540 (Envisage 7.0).

Closes #548

### Deleted files
- `envisage/egg_plugin_manager.py`, `envisage/egg_basket_plugin_manager.py`, `envisage/package_plugin_manager.py` — the three deprecated plugin manager classes
- `envisage/egg_utils.py` — utility module used only by the egg plugin managers
- `envisage/tests/test_egg_plugin_manager.py`, `envisage/tests/test_egg_basket_plugin_manager.py`, `envisage/tests/test_package_plugin_manager.py` — associated tests
- `envisage/tests/eggs/`, `envisage/tests/bad_eggs/`, `envisage/tests/plugins/` — test data directories

### Edited files
- `envisage/api.py` — removed `EggPluginManager` from public API
- `envisage/application.py` — updated comment to reference `PluginManager` instead of `EggPluginManager`
- `envisage/composite_plugin_manager.py` — updated docstring example to use `PluginManager`
- `envisage/tests/support.py` — removed egg/package-related test helpers (`restore_pkg_resources_working_set`, `build_egg`, `PLUGIN_PACKAGES`, `BAD_PLUGIN_PACKAGES`) and the now-unused `subprocess` and `pkg_resources` imports
- `pyproject.toml` — removed egg/plugin test data from `[tool.setuptools.package-data]`
- `MANIFEST.in` — removed `recursive-include` lines for deleted test data directories
- `docs/source/conf.py` — added `|PluginManager|` RST substitution
- `docs/source/envisage_core_documentation/plugins.rst` — rewrote to reference `PluginManager` instead of Python Eggs
- `docs/source/envisage_core_documentation/message_of_the_day.rst` — removed reference to egg-based plugin discovery
- `docs/source/envisage_core_documentation/extension_points.rst` — removed unused Python Eggs link
- `CHANGES.rst` — added Version 8.0.0 removal entry

### Notes
- The `setuptools` runtime dependency is retained because `pkg_resources.resource_filename` is still used in 3 other files (`test_core_plugin.py`, `test_tasks_application.py`, `_etsdemo_info.py`). Migrating those to `importlib.resources` is a separate task.
[...]

PR #612, opened 2026-08-05
Remove PluginManager include and exclude traits

### LINKED ISSUE #547: Remove PluginManager include and exclude traits
The `PluginManager.include` and `PluginManager.exclude` traits were deprecated in #544 (targeting the Envisage 7.0.0 release). For Envisage 8.0.0, we should remove the `include` and `exclude` traits, together with associated machinery and tests.

mdickinson:
Closes #547.

The `include` and `exclude` traits of `PluginManager` were deprecated in #544, targeting removal in Envisage 8.0.0.

## Changes

**`envisage/plugin_manager.py`**
- Removed the `include` and `exclude` trait declarations and the `DeprecationWarning` emitted from `__init__`.
- Removed the filtering helpers `_include_plugin`, `_is_excluded` and `_is_included`.
- `__iter__` no longer filters. It returns an iterator over a *copy* of the plugin list, which preserves the previous behaviour — the old filtered list comprehension was implicitly a snapshot (see the note below).
- `get_plugin` keeps its existing `for`/`else` structure and single trailing `return`; only the `_include_plugin` check is removed.
- Dropped the now-unused `warnings` and `fnmatch` imports, and `Str` from the `traits.api` import.

**`envisage/tests/test_plugin_manager.py`**
- Removed the four include/exclude tests and the `_test_start_and_stop` helper that only those four used. The remaining tests already cover the unfiltered behaviour.
- Added a regression test for the iteration snapshot semantics.

**`CHANGES.rst`** — added an entry under the 8.0.0 "Removals" section.

## Note on iteration semantics

The obvious simplification of `__iter__` — `return iter(self._plugins)` — would have been a silent behaviour change. The old body built a new list and iterated that, so the iteration was a snapshot; iterating the live trait list instead **silently skips** a plugin when one is removed mid-iteration, and picks up plugins added mid-iteration. This is public API: `PluginManager.__iter__` is reachable through `Application.__iter__`, and both classes document `for plugin in ...` as the supported way to iterate. The second commit restores the snapshot and adds a test that fails without it.

## Notes

These traits existed largely to filter plugins discovered by the egg- and package-scanning plugin managers. With those removed in #599, they had no remaining purpose. A grep confirms no other references anywhere in `envisage/`, `docs/`, or `examples/`.

## Testing

- Full suite passes: 179 passed, including GUI-toolkit tests.
- `black --check`, `isort --check` and `flake8` all clean.


mdickinson:
Pushed a follow-up commit (f0d234af) after review.

Collapsing `__iter__` to `iter(self._plugins)` looked like a pure simplification but changed observable behaviour. The old body built a new list before iterating, so iteration was a snapshot. Handing out an iterator over the live trait list instead means:

- removing a plugin mid-iteration **silently skips** an element (verified: iterating three plugins while removing the second visits `['a', 'c']`, no exception);
- adding a plugin mid-iteration feeds it into the in-flight iteration.

`Application.__iter__` forwards straight to this, and both classes' docstrings advertise `for plugin in ...`, so it's public API rather than an internal detail. `CompositePluginManager` is unaffected — it builds its own list first.

Restored by iterating over a copy, matching the existing `self._plugins[:]` idiom already used in `stop()`, plus a regression test that fails without the fix.

Issue #636, opened 2026-08-06
Replace black, isort and flake8 with ruff


mdickinson:
As a followup to the recent build modernization, let's replace the combination of `black`, `isort` and `flake8` with `ruff`.

PR #638, opened 2026-08-06
Reformat the codebase with ruff format

mdickinson:
Part of #636, stacked on #637. Based on `ruff-config`, so the Files tab shows
only this change; GitHub will retarget it to `main` when #637 merges.

Two commits, ordered so that each one is green:

1. **The reformat.** Mechanical — the entire diff is the output of
   `ruff format .`, touching `.py` files only. 133 files, +196/-435.
2. **One line** adding `ruff format --check --diff` to the style workflow.
   Separated from the reformat so the check is only turned on once the
   codebase already satisfies it, rather than merging something red.

## What's actually in the diff

Almost all of it is docstring normalization:

```diff
-""" An application event. """
-
+"""An application event."""
```

That is the same change black wants at its current version, so it was coming
either way — see #632. The remainder is lines rejoined at the new length of
88.

## Checks run locally

- `uv run --locked --only-group style -m ruff check .` — passes
- `uv run --locked --only-group style -m ruff format --check .` — 215 files
  already formatted
- `uv run -m pytest` — 178 passed

*This PR was created with the assistance of an AI coding agent.*


mdickinson:
Superseded by #642, which is built from scratch on current `main` rather than rebased. This branch predated the review changes on #637, so recreating it clean makes it easier to verify that the diff is exactly `ruff format` output.

PR #639, opened 2026-08-07
Create the Python shell view's control, and unwind it safely

larsoner:
@prabhuramachandran mentioned a bug to me:

> If I enable the "View->Other->Python", I get this:

```
  ETS/envisage/envisage/plugins/python_shell/view/python_shell_view.py", line 146, in destroy_control
    self.window.application.unregister_service(self._service_id)
                                               ^^^^^^^^^^^^^^^^
AttributeError: 'PythonShellView' object has no attribute '_service_id'
```

I had Claude Opus 5 look into it and below was the conclusion it wrote out. Happy to iterate on the PR and own/vouch for these changes, even if I'm a little out of my depth a bit here -- the associated workaround in https://github.com/enthought/mayavi/pull/1414 does seem to work!

---
 
Pyface 8.0 split widget construction in two: PythonShell(parent) leaves 'control' at None until create() is called.  'create_control' never made that second call, so it went on to read 'self.namespace' off a shell with no interpreter and died with

    AttributeError: 'NoneType' object has no attribute 'interpreter'

which is to say the view has not opened at all under any Pyface since 8.0. Call create() when the control is not already there, so this keeps working against a Pyface that still builds it in the constructor.

What the user actually sees is a different error, from the clean-up:

    AttributeError: 'PythonShellView' object has no attribute '_service_id'

'create_control' assigns '_service_id' on its last line, and Pyface's WorkbenchWindowLayout.add_view calls 'destroy_control' from the except branch that handles a 'create_control' which raised -- so undoing what never happened replaced the error explaining the failure with an error from the clean-up after it.  Give '_service_id' a trait default and skip what was never done; same for the sys.stdout redirect, which a part-way 'create_control' does leave behind.

These are the plugin's first tests, and the first anywhere in envisage to import pyface.workbench, whose import-time PendingDeprecationWarning the pytest configuration has to allow now that warnings are errors.

Reported downstream as enthought/mayavi#1409.


mdickinson:
Thanks, @larsoner! Looks fine in principle. A few small things:

- it's fine for the new test-subpackage `__init__.py` to be empty - no need for a copyright header in what would otherwise be an empty file
- EDM tests are failing because `PythonShell` depends on `pygments`. The non-EDM tests are passing only accidentally, because the version of `pytest` used for those tests pulls in `pygments`.
- Claude pointed out that the original `stdout` might genuinely be `None`, in which case the restoration in `destroy_control` would now be bypassed. Not sure what the cleanest way to fix that is. Claude suggested an extra boolean flag.
- Copilot has a reasonable point about cleanup - it would be good to be sure that the new tests can't leave global state changed (on either pass or failure).

Let me see if I can whip up a quick fix for the `pygments` dependency in a separate PR.

Also worth noting: I'm about to merge #638, which _might_ cause this PR to need a reformat pass.

mdickinson:
> Let me see if I can whip up a quick fix for the `pygments` dependency in a separate PR.

#641 (merged) ensures that pygments is present in the test environment for the GUI-using tests, and that a test marked with `@requires_gui` is skipped if pygments isn't available.

There's a catch, though - for the one non-gui-using test job, the new test module _may_ still fail at import time, in which case we'd need a local import.

mdickinson:
> I'm about to merge #638, which _might_ cause this PR to need a reformat pass.

Looks like it did. Sorry about that. :-(

larsoner:
Done @mdickinson !

mdickinson (approved):

mdickinson:
Thanks, @larsoner! Getting a second local review from Opus, then assuming that turns up nothing blocking I'll merge.

mdickinson:
Addressed my own `__init__.py` nitpick in [9faadbb](https://github.com/enthought/envisage/pull/639/commits/9faadbb89ba93088a3799df4a151c39b47d573c1). I do still want to take a second look at the `original_stdout` issue; doing that shortly.

mdickinson:
@larsoner This fix has now been shipped in Envisage 8.0.1.

larsoner:
Awesome, thanks!

prabhuramachandran:
Many thanks @larsoner and @mdickinson! Do you have an idea of when this would be in a new envisage release? Is there a planned release in the near future? Thanks.

prabhuramachandran:
@mdickinson -- pardon the noise, I just saw the release!

PR #654, opened 2026-08-07
Prepare the 8.0.1 release

### LINKED ISSUE #653: Release Envisage 8.0.1
A bugfix release, made directly from `main`: there are no non-bugfix
user-facing changes since 8.0.0, so no release branch is needed.

What it contains:

* The Python shell view opens again. It has not opened under any Pyface
  since 8.0, which split widget construction in two (#639, reported
  downstream as enthought/mayavi#1409).
* The Read the Docs build works again, and several smaller problems in the
  generated documentation are fixed (#645, #648, #652).

The rest of what has landed since 8.0.0 is build and test machinery, which
the changelog records but which doesn't affect users of the package.

Steps:

- [ ] Set the release date and version in `CHANGES.rst`, bump `pyproject.toml`, refresh `uv.lock`
- [ ] Tag the merge commit, with the release notes as the tag message
- [ ] Create the GitHub release as a draft
- [ ] Publish the release, which triggers the PyPI upload and the docs deployment

mdickinson:
Closes #653.

Prepares the 8.0.1 release:

* Retitle the in-development changelog section from `8.1.0` to `8.0.1`, and
  replace the `Released: YYYY-MM-DD` placeholder with today's date.
* Add the overview, in the style used for earlier bugfix releases, and credit
  contributors since 8.0.0 under `Thanks to`.
* Bump the version in `pyproject.toml` from 8.0.0 to 8.0.1, and refresh
  `uv.lock`, which pins the project's own version. (8.0.0 predated the
  lockfile, so this step is new since the last release.)

Three PRs merged since the 8.0.0 tag had no changelog entry, so this adds
them: the ruff reformat (#642), folded into the entry for the PR that
configured ruff (#637), the `edm.yaml` move (#634), and the
`pypa/gh-action-pypi-publish` bump (#633). #650 is deliberately left out,
since it only edited the changelog. Happy to drop #633 and #634 if you'd
rather keep the release notes to things a reader of them would act on.

Pre-flight, beyond CI: `uv build` and `twine check --strict` both pass, and
`uv sync --locked` confirms the refreshed lockfile.

Diffing the 8.0.1 artifacts against the released 8.0.0 ones shows one
difference that isn't explained by the changes since 8.0.0: the sdist now
carries a `pyproject.toml.orig` beside a `pyproject.toml` that has been
rewritten and stripped of its comments. That is `uv_build` working as
designed — since uv 0.12.0 it always rewrites the sdist's `pyproject.toml`
as TOML 1.0, so that older build frontends can consume it, and preserves the
original alongside. The two parse to identical data. It arrived with #628,
along with the switch to `uv_build`, and needs no action.

*This PR was created with the assistance of an AI coding agent.*

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -2,10 +2,24 @@
  Release history
 =================
 
-Version 8.1.0
+Version 8.0.1
 =============
 
-Released: YYYY-MM-DD
+Released: 2026-08-07
+
+This is a bugfix release aimed at making the Python shell view usable again.
+A change in the way Pyface constructs widgets stopped the view opening at
+all, so ``View -> Other... -> Python`` failed in Envisage-based applications
+such as Mayavi.
+
+The release also restores the Envisage documentation on Read the Docs, whose
+builds had been failing, and fixes several smaller problems in the generated
+documentation.
+
+Thanks to:
+
+* Mark Dickinson
+* Eric Larson
 
 Fixes
 -----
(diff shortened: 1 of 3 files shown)


mdickinson:
One thing on the overview, which otherwise reads well and is right to lead with
the shell view.

The move to the `uv_build` backend is the change in this release with the best
chance of biting someone, and at the moment it appears only as a Build bullet.
8.0.1 is the first release whose source distribution declares `uv_build` in
place of `setuptools`, in a `src/` layout and without `CHANGES.rst`, so anything
that builds Envisage from the sdist without build isolation needs its own build
requirements updated to match.

That isn't hypothetical: the conda-forge recipe builds with `pip install . -vv
--no-deps --no-build-isolation` and lists `setuptools`, `wheel` and `pip` as
host requirements, so it will need `uv-build` there before it can take 8.0.1.
`uv-build` 0.12.2 is already on conda-forge, so that's a one-line change for
them rather than a blocker — but it is a surprising thing to meet in a patch
release, and a reader who stops at the end of the overview, which is most of
them, won't see it coming.

Suggest a sentence at the end of the overview, something like:

> Building from the source distribution now requires the ``uv_build`` backend
> in place of ``setuptools``; installing from the wheel, as ``pip install
> envisage`` does, is unaffected.

